PR #3, opened 2017-05-26
The provided version of pear doesn't work anymore in UB 16.04, some…

tschoening:
…CTOR call with "new" somewhere doesn't work anymore. Simply deleting PEAR and using that from the distribution instead works as needed.

> apt-get install php-pear
> pear install Archive_Tar
> pear install Text_Diff


joubertredrat:
Thanks again with your contribution @tschoening, merged.

PR #4, opened 2017-08-24
"showDirFiles" might create unnecessary null entries in $listing.

tschoening:
"showDirFiles" always created a new entry in the structure containig the dirs and files to render, even if some dir was not accessible for the current user because a lack of permissions. While those entries were null/undef or such, during rendering the template the data of the last non-null entry was used to render all the null-entries as well. This resulted in wrong HTML output of course.


joubertredrat:
Thanks for the contribution @tschoening. Merged

PR #9, opened 2018-02-14
Adding basic docs about required PEAR-Libs.

tschoening:
The PEAR-dependencies Archive_Tar and Text_Diff have been removed already, but I didn't provide any docs how to install those. It's only documented in the PR, which is not very user-friendly. This commit adds some basic docs, especially because Geshi gets removed in the near future as well and can easily be added there afterwards.

Additionally an unnecessary closing "div" is removed, because I don't think this little change to make the HTML document valid is necessary another branch/commit.

### DIFF
--- a/doc/install.html
+++ b/doc/install.html
@@ -59,7 +59,13 @@
 
 <h2><a name="installation"></a>Installation</h2>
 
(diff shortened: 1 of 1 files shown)


tschoening:
Didn't change anything on the required version for PHP and SVN for now because of #8. During that work Geshi needs to be added and the version of SVN improved because of svnauthz etc. anyway.

PR #54, opened 2018-07-27
Remove PHP Warning when no extension for file

tonvoon:
This removes the warning message:

PHP Warning:  array_key_exists(): The first argument should be
either a string or an integer in include/svnhook.php on line 643


ams-tschoening:
Thanks for fixing!

PR #66, opened 2018-12-08
Use ignoreWhitespace option in compare page

Tom-H-:
Use the `ignoreWhitespace` option in the compare page as in the diff page.


ams-tschoening:
Thanks for the contribution!

PR #81, opened 2019-05-10
websvn crashes if the user has access to no repositories

trentfisher:
If a user has access to no repositories, it crashes with this error:

Fatal error: Uncaught Error: Call to a member function hasUsername() on null in /scm/share/apache2/htdocs/websvn-2.5/include/setup.php:568
Stack trace:
#0 /scm/share/apache2/htdocs/websvn-2.5/index.php(107): checkSendingAuthHeader()
#1 {main} thrown in /scm/share/apache2/htdocs/websvn-2.5/include/setup.php on line 568

This patch prevents the crash.

### DIFF
--- a/include/setup.php
+++ b/include/setup.php
@@ -565,6 +565,7 @@ function checkSendingAuthHeader($rep = false) {
 	} else {
 		$authz =& $config->getAuthz();
 	}
-	$loggedin = $authz->hasUsername();
+        // authz can be null if the user has access to no repositories
+	$loggedin = ($authz && $authz->hasUsername());
 	http_response_code(403);
 }


ams-tschoening:
> // authz can be null if the user has access to no repositories

Are you sure about that, did you really configure an access file? From my understanding `authz` is only `null` if no access file has been configured at all or can't be found. Else how can WebSVN know that the current user doesn't have access at all?

I don't understand the purpose of that function at all: `$loggedin` isn't used anywhere, but the 403-header is sent unconditionally. It seems the callers already checked if someone has access, so that function doesn't need to anymore. Seems to me the correct patch would be removing the check altogether and only sending the header. Renaming the function to something like `sendUnauthedHeader` or something might make sense as well.

ams-tschoening:
The check has been disabled by purpose in the following commit:

https://github.com/websvnphp/websvn/commit/0aaa98b6f9562b7cea77d9b38ea7ba3718956482#diff-4e85e5ab5c8d90e0db882ad816cbba5c

The following line is of special interest:

    header('WWW-Authenticate: Basic realm="'.str_replace('"', '\"', $auth->getBasicRealm()).'"');

So the ultimate goal was to let users authenticate as needed, using some pieces of information being formerly available in `auth` and that object being available always. That has changed in the commit replacing custom authz-parsing with using `svnauthz`:

https://github.com/websvnphp/websvn/commit/9b42ab9ce47011c54d1c3a32395120b5102559b1

But a former commit already removed basic auth handling altogether, leaving that to the web server, to make the code base easier:

https://github.com/websvnphp/websvn/commit/c1b75d66da7d67502854c9b20efa75e056d1dd7b#diff-4e85e5ab5c8d90e0db882ad816cbba5c

So, as already guessed, the purpose of this function simply doesn't exist anymore and the proper fix is to remove it. I will do that and close this PR therefore.

Thanks for pointing at this problem!

PR #84, opened 2019-05-24
checkSendingAuthHeader is of no use anymore.

ams-tschoening:
Since handling basic auth has been removed entirely and the main logic has been commented because of problems anyway already. If it only sends a 403-header these days, all other checks are unnecessary and make unnecessary problems, so I've removed them and renamed the function.

https://github.com/websvnphp/websvn/pull/81

### DIFF
--- a/blame.php
+++ b/blame.php
@@ -211,7 +211,7 @@ if ($rep) {
 
 	if (!$rep->hasReadAccess($path, false)) {
 		$vars['error'] = $lang['NOACCESS'];
-		checkSendingAuthHeader($rep);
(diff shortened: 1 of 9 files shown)


ams-tschoening:
> Does it make sense to have this function at all if it only does: http_response_code(403)?

Wasn't sure about that as well, but this function only sets the header, it doesn't e.g. end the request at all. So what if one wants to end the request instantly in future using an additional `exit` or stuff? What if additional template processing should happen to localize messages or else?

In the end, I thought it was easier to leave the function in place for now instead of adding it back if one changes its mind.

Issue #104, opened 2020-05-21
In Listing View Cannot Open Two Directories in a Tree View for Comparison


k10blogger:
One of the most commonly used feature atleast for me is the compare paths in the listing view. 

The issue is, it is not possible in the listing view to compare same file across multiple tags or branches or even in different folders in a same tag. 

For example consider below two directories under tags(as seen in image).
[1] and [2] as in image attached.
As [2] is open i can see files within it. But if i have to compare it with files in [1] I cannot use this UI for comparison as when i open the folder [1] the folder [2] closes.

![image](https://user-images.githubusercontent.com/4973182/82509255-f46af680-9b24-11ea-9e3e-d55a1b68f524.png)

Things are not better when i select a file in tag and try to compare it with one in trunk. As soon as i click on the trunk folder the tag folder closes. 

Anyway of fixing this?

ams-tschoening:
I don' think so, WebSVN is a pretty old-school web app requesting one path only and NOT storing any former state AFAICS. The latter would be what you need to compare arbitrary paths. Looking a bit into things, here is how I think showing the current directory works:

Consider a path `/foo/bar` requested, than contents of `/`, `/foo` and `/foo/bar` get build and rendered, in the order. What you would need to have is somehow remembering the path your are interested in besides the current directory and render that pretty much the same way like the current directory. Afterwards providing all those pieces to the template engine.

Besides thinking of some way to have the former path to compare with still available, this requires heavy refactoring of `listing.php`. It assumes to be called with the current directory only and relies heavily on globally available variables.

I'm attaching some PoC of how things need to be changed, but don't have the time to work more on this. What's currently especially not working is rendering contents for each path, because some index is used and most likely places things at the same places at some point or such. When trying two different paths, the template show some parts doubled and some parts of the latter of the two paths only.

[listing_with_two_paths.patch.txt](https://github.com/websvnphp/websvn/files/4661886/listing_with_two_paths.patch.txt)
[listing.php.txt](https://github.com/websvnphp/websvn/files/4661887/listing.php.txt)

k10blogger:
Hi I have been working on this but not had much success. I first learned php for 10-15 days but whatever i learnt had no implementation here. 

I implemented what you mentioned [here](https://github.com/k10blogger/websvn/commit/76d25be68efe37188740b57fa5a037448277aa31). This can be expanded for n number of path. 

[...]

PR #111, opened 2020-08-05
Collapsible Files in Compare View

### LINKED ISSUE #108: Collapsible Files in Compare View
The expectation is to have a collapsible file comparison view. Here in the snapshot below is an attempted slight improvement in the existing compare view.
image
1 - Is the view of collapsed file. You click on the file bar and it collapse
2 - The expanded view click on view and it expands.
I just used the existing jquery functions to achieve this and here is the commit snapshot
image

It is really helpful in scenarios when comparing two folders in subversion with lot of files in them at different revisions. 
I didnt see any use of jquery as it is purely in php so not sure this is an expected way of implementing this.
Any thoughts on this?

k10blogger:
The commits are in to perform the collapsible file operation in the compare view. Fixes #108 
Following changes have been done:
1 - Renamed the group-collapse.js to collapsible.js
2 - Renamed the references of group-collapse.js to collapsible.js
3 - Add the collapsefile() to the collapsible.js (Possibly an appropriate function name. Name refactoring can be done here.)
4 - Removed refernces to the opentree and collapseAllGroups

Could not do icons.


k10blogger:
After creating this pull request i realized that i have commits for the #104 .
I have now removed the test code but kept the changes that were suggested for the same. 
This has no impact on the actual operation.

listing.php
```diff
@@ -206,12 +205,14 @@ function showTreeDir($svnrep, $path, $rev, $peg, $listing) {
 	}
 
 	$vars['compare_box'] = ''; // Set blank once in case tree view is not enabled.
-	return showDirFiles($svnrep, $subs, 0, $limit, $rev, $peg, $listing, 0, $config->treeView);
+	//return showDirFiles($svnrep, $subs, 0, $limit, $rev, $peg, $listing, 0, $config->treeView);
```
michael-o, line 208: No need for such comments, we have source control.

listing.php
```diff
@@ -333,15 +334,28 @@ function showTreeDir($svnrep, $path, $rev, $peg, $listing) {
 	$vars['compare_endform'] = '</form>';
 
 	$vars['showlastmod'] = $config->showLastModInListing();
-
-	$listing = showTreeDir($svnrep, $path, $rev, $peg, array());
-
+	//$listing = showTreeDir($svnrep, $path, $rev, $peg, array());
+	$listing = showTreeDir($svnrep, $path, $rev, $peg, $listing,$index);
```
michael-o, line 338: Same here.

javascript/collapsible.js
```diff
@@ -18,3 +18,11 @@ function collapseAllGroups() {
       tableRows[i].style.display = 'none';
   }
 }
+function collapsefile(CurrObj) {
```
ams-tschoening, line 21: In general, just follow camel case like the other code uses : `collapsefile` -> `collapseFile`, `CurrObj` -> `currObj`

BUT: I don't like `onclick` and calling this function in the HTML. Why did you introduce it at all? Your solution with registering a click-handler in JS afterwards looks far better. Why is that commented, didn't work? I prefer those, not `onclick`.

javascript/collapsible.js
```diff
@@ -18,3 +18,11 @@ function collapseAllGroups() {
...
+    CurrObj.closest("table").find("tbody").toggle(); 
+}
+/*
+$("thead").find("th").on("click", function() {
```
ams-tschoening, line 25: This is the preferred solution, because it doesn't require HTML-changes.

templates/BlueGrey/compare.tmpl
```diff
@@ -36,7 +36,7 @@
       [websvn-test:newpath]
       <table class="comparison">
         <thead>
-        <tr><th><a href="[websvn:fileurl]">[websvn:newpath]</a></th></tr>
+        <tr><th onclick="collapsefile($(this));"><a href="[websvn:fileurl]">[websvn:newpath]</a></th></tr>
```
ams-tschoening, line 39: Remove `onclick` in favour of registering handlers for that event in `collapsible.js` only. That's the far better solution and you already had it! :-) If that doesn't work, let's discuss the reason and try to fix it.

include/setup.php
```diff
@@ -493,6 +493,15 @@
 $path = !empty($_REQUEST['path']) ? escape($_REQUEST['path']) : null;
 if ($path === null || $path === '')
 	$path = '/';
+if(isset($_COOKIE['prevpath']))
```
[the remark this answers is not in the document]
k10blogger: I have reverted all the changes related to #104.

listing.php
```diff
@@ -107,9 +107,8 @@ function showDirFiles($svnrep, $subs, $level, $limit, $rev, $peg, $listing, $ind
 			$isDirString = ($isDir) ? 'isdir=1&amp;' : '';
 
 			// Only list files/directories that are not designated as off-limits
-			$access = ($isDir)	? $rep->hasReadAccess($path.$file, false)
-								: $accessToThisDir;
-
+			$access = ($isDir) ? $rep->hasReadAccess($path.$file, true)
+												 : $accessToThisDir;
```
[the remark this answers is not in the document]
k10blogger: I have reverted it back.

listing.php
```diff
@@ -191,7 +190,7 @@ function showDirFiles($svnrep, $subs, $level, $limit, $rev, $peg, $listing, $ind
 	return $listing;
 }
 
-function showTreeDir($svnrep, $path, $rev, $peg, $listing) {
+function showTreeDir($svnrep, $path, $rev, $peg, $listing, $index) {
```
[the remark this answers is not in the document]
k10blogger: Removed.

listing.php
```diff
@@ -206,12 +205,14 @@ function showTreeDir($svnrep, $path, $rev, $peg, $listing) {
...
+	
+	return showDirFiles($svnrep, $subs, 0, $limit, $rev, $peg, $listing, $index, $config->treeView);
 }
 
-// Make sure that we have a repository
-if ($rep) {
-	$svnrep = new SVNRepository($rep);
+function procPath($svnrep, $path, $listing, $index)
+{
+	global $vars, $config, $lang, $rep, $repname, $rev, $passrev, $peg;
```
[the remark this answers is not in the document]
k10blogger: Removed.

listing.php
```diff
@@ -333,15 +334,35 @@ function showTreeDir($svnrep, $path, $rev, $peg, $listing) {
 	$vars['compare_endform'] = '</form>';
 
 	$vars['showlastmod'] = $config->showLastModInListing();
-
-	$listing = showTreeDir($svnrep, $path, $rev, $peg, array());
-
+	//$listing = showTreeDir($svnrep, $path, $rev, $peg, array());
+	$listing = showTreeDir($svnrep, $path, $rev, $peg, $listing,$index);
```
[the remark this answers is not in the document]
k10blogger: removed.

listing.php
```diff
@@ -333,15 +334,35 @@ function showTreeDir($svnrep, $path, $rev, $peg, $listing) {
...
 	if (!$rep->hasReadAccess($path)) {
 		$vars['error'] = $lang['NOACCESS'];
 		sendHeaderForbidden();
 	}
 	$vars['restricted'] = !$rep->hasReadAccess($path, false);
 
+	return $listing;
```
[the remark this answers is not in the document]
k10blogger: removed.

javascript/collapsible.js
k10blogger: They were two though process for that. You had requested not to have it inline. I though this what you meant. That is my mistake. The other was i wanted to use the pre-existing collapsing icons that were there in the templates. I wanted to reuse them, Thought that was the reason you asked to not make it inline. 
I do not know how to work with the CSS or add icons. I thought of reusing what is already there.

templates/BlueGrey/compare.tmpl
```diff
@@ -58,4 +58,6 @@
       [websvn-endtest]
     [websvn-endlisting]
   [websvn-endtest]
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>
```
michael-o, line 61: Not good, maybe not ideal in corporate environment where deployments must be selfcontained.

javascript/collapsible.js
k10blogger: Have a look at this 
image
This has a class name open and closed. So when it is collapsed it refers to "th.closed" and when it is expanded it refers to "th.open". This is in CSS file for the template. This is common for all templates. Unfortunately all templates do not have toggle icons like calm.

templates/BlueGrey/compare.tmpl
k10blogger: We can download the version and include it in the javascript folder. That is what i have done in my organization.

michael-o: This is prefered.

listing.php
k10blogger: Removed as these were the changes for #104

listing.php
k10blogger: Removed as these were the changes for #104

javascript/collapsible.js
k10blogger: With the new commit I have removed the collapsefile and kept the original solution :). Will think of applying the icon in some different way.

javascript/collapsible.js
k10blogger: Yes with the new commit made the preferred solution. Revert the HTML Changes as well.

templates/BlueGrey/compare.tmpl
k10blogger: Yes i thought by inline you meant something like onclick style function that is there already there and then move the function to separate JS file. :) No issues i have revert this HTML changes and went back to the original solution.

templates/BlueGrey/compare.tmpl
k10blogger: Done.

ams-tschoening: Bundling jQuery the current way is WRONG! Because a revision number or stuff like that is missing, one can't influence anymore if an ever updated version of jQuery will be downloaded by the browser or not. The former implementation referencing the file from some CDN was the better and more standard approach. A lot of web-apps do so and because of that, chances are high that some browser already has the requested version of jQuery in its own cache. This even reduces load on the corporate IT.

@michael-o Which problem do you hope to solve with bundling jQuery? We really should NOT do so.

javascript/collapsible.js
```diff
@@ -18,3 +18,7 @@ function collapseAllGroups() {
       tableRows[i].style.display = 'none';
   }
 }
+$("thead").find("th").on("click", function() {
```
ams-tschoening, line 21: This code registers toggle-callbacks for all pages which load `collapsible.js` right now. Is that as expected and does it work for all tables on pages not related to comparing revisions? If it works and doesn't break anything else, let's keep it. If it doesn't work, the tables you want to collapse should get an additional CSS-class like `collapsible` and `$("thead")` should become something like `$("table.collapsible thead")` instead.

templates/BlueGrey/index.tmpl
```diff
@@ -55,14 +55,6 @@
   [websvn-endtest]
     </table>
   [websvn-test:treeview]
-    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/group-collapse.js"></script>
-    [websvn-test:opentree]
```
ams-tschoening, line 59: Are you sure removing those parts is correct? I don't see any other implementation of that behaviour in JS. Looks to me like this needs to be reintroduced.

templates/Elegant/compare.tmpl
```diff
@@ -38,7 +38,7 @@
       [websvn-test:newpath]
       <table>
         <thead>
-        <tr><th><a href="[websvn:fileurl]">[websvn:newpath]</a></th></tr>
```
ams-tschoening, line 41: Removing this link looks like an error.

templates/Elegant/index.tmpl
```diff
@@ -71,14 +71,6 @@
   [websvn-endtest]
     </table>
   [websvn-test:treeview]
-    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/group-collapse.js"></script>
-    [websvn-test:opentree]
```
ams-tschoening, line 75: Same like before, you seemed to have removed some feature.

templates/calm/compare.tmpl
```diff
@@ -46,7 +46,7 @@
     [websvn-test:newpath]
     <table class="diff">
       <thead>
-      <tr><th><a href="[websvn:fileurl]">[websvn:newpath]</a></th></tr>
```
ams-tschoening, line 49: Missing link again.

templates/BlueGrey/compare.tmpl
michael-o: @ams-tschoening Some people dislike serving resources from outside their network. There is no real problem here.

templates/calm/index.tmpl
```diff
@@ -59,14 +59,6 @@
     </table>
   </div>
   [websvn-test:treeview]
-    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/group-collapse.js"></script>
-    [websvn-test:opentree]
```
ams-tschoening, line 63: Same like before, you seemed to have removed some feature.

templates/BlueGrey/index.tmpl
```diff
@@ -55,14 +55,6 @@
...
-    [websvn-else]
-    <script type="text/javascript">
-    //<![CDATA[
-    collapseAllGroups();
-    //]]>
-    </script>
-    [websvn-endtest]
+    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>
```
ams-tschoening, line 58: You can't just use the one line for `collapsible.js` only, because that file contains code which uses jQuery always. You always need both lines:

    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/jquery.min.js"></script>
    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>

templates/BlueGrey/compare.tmpl
ams-tschoening: Those things can be addressed when having a proper JS-build with package management, cache busting etc. in place. We shouldn't care now unless anyone has a problem, which in my opinion is pretty unlikely. Bundling like currently makes the app only less reliable and means more work for us when jQuery needs to be updated for some reason. Lot's of apps don't use jQuery at all anymore as well.

I strongly suggest using the following line again:

    <script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>

javascript/collapsible.js
ams-tschoening: If icons are available already, things are pretty easy: Change the HTML-template to ALWAYS add the "open" class. This is [done in some templates already](https://github.com/websvnphp/websvn/blob/master/templates/calm/index.tmpl#L30):

      [websvn-test:opentree]
              class="open"
      [websvn-else]
              class="closed"
      [websvn-endtest]

Afterwards you change usage of `toggle` in your JS to [toggleClass](https://api.jquery.com/toggleclass/#toggleClass-function-state). This way you can check if the currently applied class is `open` or `close` and change accordingly. Should be something like the following:

    $(this) .closest("table").find("tbody")
            .toggleClass(function(idx, oldClass, state)
    {
      return oldClass == "open" ? "closed" : "open";
    });

You need to get syntax and interaction of `oldClass` vs. `state` correct on your own.

templates/BlueGrey/compare.tmpl
k10blogger: Probably @michael-o is referring to the corporate networks that are isolated. Like I have in our lab. They cannot go outside the firewall and hence will never be able to fetch this js file. These networks are strictly for source code maintenance and testing. Maybe to support those this would makes sense to bundle together. 
But again with more and more corporate outsourcing the source code servers to cloud maybe there is no need to bundle it with this websvn code. 
I am fine with either. But i would prefer this
`<script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>`
Let me know which one to do.

ams-tschoening: If the external file works for you, please use the external one. If not, bundle it, but at least renamed to contain the version number, like `jquery.min.1.9.1.js` or something.

javascript/collapsible.js
k10blogger: This only works on the comparison revision. It doesnt seem to work on anything else. The reason might be that the comparison view is using a different css style than the rest of the templates.

templates/BlueGrey/compare.tmpl
k10blogger: I have bundled it for now as in my lab we dont have a network connection so the external one is not working. I have bundled it and renamed it to jquery.min.1.9.1.js so that the name version is clear. :-)

templates/BlueGrey/index.tmpl
k10blogger: No No no this was not correct i tried to move this ope-tree selection in JS but couldnt succeed. So I have reverted it back. This i will take care in separate issue. For now i have reverted it back with collapsible.js and added the call of jquery-min version 1.9.1 file on top of this line.

templates/Elegant/compare.tmpl
k10blogger: Fixed with the new commit.

templates/Elegant/index.tmpl
k10blogger: i tried to move this ope-tree selection in JS but couldnt succeed. So I have reverted it back. This i will contribute or take care in separate issue. For now i have reverted it back with collapsible.js and added the call of jquery-min version 1.9.1 file on top of this line.

templates/calm/compare.tmpl
k10blogger: Fixed with the new commit.

templates/calm/index.tmpl
k10blogger: i tried to move this ope-tree selection in JS but couldnt succeed. So I have reverted it back. This i will contribute or take care in separate issue. For now i have reverted it back with collapsible.js and added the call of jquery-min version 1.9.1 file on top of this line.

templates/BlueGrey/index.tmpl
k10blogger: I have added the 
```
    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/jquery.min.1.9.1.js"></script>
    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>
```
in all the three index.tmpl

javascript/collapsible.js
k10blogger: I have successfully done for the calm template but for the other two there are issues.
I implemented in a sligthly different way the code is below:
```
$("thead").find("th").on("click", function() {
    $(this).get(0).className = ($(this).get(0).className == 'open') ? 'closed' : 'open';
    $(this).closest("table").find("tbody").toggle();
    
});
```
I defaulted the class to open in the html template.
Then i added the following in the CSS
```
thead th.open {
	background-image:url(images/toggledown.png);
	background-repeat:no-repeat;
}
thead th.closed {
	background-image:url(images/toggleup.png);
	background-repeat:no-repeat;
}
```
This is how it looks in calm template
image

This all is in the commit.
This same i could not apply for the other two templates.
For elegant the comparison tables uses a repeated image to generate the background
image
For bluegrey there is no template specific icon present. If you can provide some solution for those two after looking in the commit i can do that change as well.

michael-o:
@k10blogger Can you provide two screenshots, before and after?

javascript/collapsible.js
ams-tschoening: Please rename the toggle-icons to `toggle-down|up.png`, as that is better readable.

If you are not interested in the other templates yourself, you might simply ignore them for now. In theory, different templates simply can look differently, so the approach with icons for `calm` might not apply to other templates.

If you care about those templates, have a look at the download-icon of `Elegant`:

https://github.com/websvnphp/websvn/blob/master/templates/Elegant/images/download.png

That might be used for both templates lacking toggle-icons, as the color scheme somewhat fits both. Simply copy and rename the file to follow the naming scheme `toggle-down|up.png` and rotate the image to get the missing toggle-icon.

As for the background of `Elegant`, I would try to create a new background image simply containing the toggle-icons and apply that wherever you want to toggle: `bg-table-header-toggle-down|up.png`, `bg-table-divider-toggle-down|up.png`

javascript/collapsible.js
ams-tschoening: But you are not checking any CSS-style when applying the click-handler, aren't you? :-) You only test for HTML-elements and `thead` and `th` are available in multiple tables. The template for `BlueGrey` contains the following HTML:

    <table class="comparison">

So please add a CSS class for those tables which should be toggled like I suggested before:

    <table class="comparison collapsible">
    $("table.collapsible thead")[...]

This makes sure that we don't interfere with unrelated tables by accident and your changes can be used for other tables as well.

k10blogger: Ahh okay now i understand. Yes this absolutely makes sense. I have added the collapsible class to each table where JS is to be applied. In all the three templates. Also modified the JS now to trigger when class of "collapsible" is clicked. :)

javascript/collapsible.js
k10blogger: I have done the changes for the calm template and provided the commit for it. 
There seems to be a problem with the BlueGrey Template. I have checked in the master copy in git(Not mine). Whenever i do compare in bluegrey it shows an error as below. This error is not there in any other template. 
The error snapshot
image
The working snapshot
image

Shall i raise an issue for this?

k10blogger:
@michael-o 
Here are the screen shots template wise.

Calm
Before:
Before-Calm
After:
After-Calm
After All Collapsed:
After-Calm-Collapsed

Elegant:
Before:
Elegant-Before
After:
Elegant-After-Collapse

javascript/collapsible.js
k10blogger: I have raised an issue for this.

ams-tschoening:
Should the up-arrow be rotated? When I look at my text editor, the icon looks different. Is the up-arrow standard this way?

Clipboard01

k10blogger:
When repositories are group the orientation is the same as committed. 
We can rotate. I dont see any problem in rotating. Maybe it will be more intuitive that way? 
What do you suggest? I am fine with either.

ams-tschoening:
Let's leave it as-as right now if it's more consistent with other places. Things can easily be changed as necessary. Is there still something open right now or can this be merged now?

k10blogger:
No nothing is open as of now.

ams-tschoening:
Thanks for the work! If anybody encounters problems, feel free to open issues.

k10blogger:
Happy to help :)

PR #113, opened 2020-08-14
Normalizing the templates for warning and changin HEAD:HEAD to HEAD:0

k10blogger:
This PR is for discussion in issue #112 
The bluegrey template has a warning block in the html template whereas the other templates didnt have that.
This PR adds the warning block to templates of calm and elegant.

There has been a discussion.

> The problem is -r HEAD:HEAD and should most likely be -r HEAD:0 instead. This is related to #102.
> 
> It's not a keyword, but I think it's what you want:
> «svn log -r HEAD:0 --limit=1 -- ${TARGET}@Head»
> 

https://mail-archives.apache.org/mod_mbox/subversion-users/202005.mbox/%3C20200524220314.1a68eb8a@tarpaulin.shahaf.local2%3E

Keeping this mailing discussion in mind it was identified that 
`function getLog($path, $brev = '', $erev = 1, $quiet = false, $limit = 2, $peg = '', $verbose = false) {`

the end revision argument in the getLog function be set to zero.

### DIFF
--- a/templates/Elegant/compare.tmpl
+++ b/templates/Elegant/compare.tmpl
@@ -36,7 +36,6 @@
     [websvn-test:warning]
       <div id="warning">[websvn:warning]</div>
     [websvn-endtest]
-    [websvn-test:success]
     <div id="comparisons">
     [websvn-startlisting]
       [websvn-test:newpath]
@@ -66,5 +65,4 @@
     </div>
    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/jquery.min.1.9.1.js"></script>
    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>
-  [websvn-endtest]
 [websvn-endtest]
--- a/templates/Elegant/compare.tmpl
+++ b/templates/Elegant/compare.tmpl
@@ -35,7 +35,8 @@
     </div>
     [websvn-test:warning]
       <div id="warning">[websvn:warning]</div>
-    [websvn-else]
+    [websvn-endtest]
+    [websvn-test:success]
     <div id="comparisons">
     [websvn-startlisting]
       [websvn-test:newpath]
--- a/templates/Elegant/compare.tmpl
+++ b/templates/Elegant/compare.tmpl
@@ -33,6 +33,9 @@
     </table>
     [websvn:compare_endform]
     </div>
+    [websvn-test:warning]
+      <div id="warning">[websvn:warning]</div>
+    [websvn-else]
     <div id="comparisons">
     [websvn-startlisting]
       [websvn-test:newpath]
@@ -63,3 +66,4 @@
    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/jquery.min.1.9.1.js"></script>
    <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>
   [websvn-endtest]
+[websvn-endtest]
--- a/comp.php
+++ b/comp.php
@@ -127,12 +127,12 @@ if ($rep) {
 	$vars['rev1'] = $rev1;
 	$vars['rev2'] = $rev2;
 
-	$history1 = $svnrep->getLog($path1, $rev1, $rev1, false, 1);
+	$history1 = $svnrep->getLog($path1, $rev1, 0, false, 1);
 	if (!$history1) {
 		http_response_code(404);
 		$vars['error'] = $lang['NOPATH'];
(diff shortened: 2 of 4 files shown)


templates/Elegant/compare.tmpl
```diff
@@ -33,6 +33,9 @@
     </table>
     [websvn:compare_endform]
     </div>
+    [websvn-test:warning]
+      <div id="warning">[websvn:warning]</div>
+    [websvn-else]
```
ams-tschoening, line 38: I suggest removing that ELSE-stuff from ALL templates: If there's a warning, it's of course fine to print that. But that shouldn't prevent any other valid output. Remember that the templates you just changed DIDN'T print the warning, but only the OTHER content and nobody cared, because things seemed to have worked. :-) By introducing the ELSE, this gets changed to NEVER print the content even if the warning isn't too important at all.

michael-o: @ams-tschoening So you are proposing that the warning should be printed additionally to the regular output?

k10blogger: With this particular change i think warning will only be generated in case the revision number is out of bounds. 
Meaning we try to compare some location or file where both revision entered are out of bounds. 

In all other cases there will be either a difference that will be shown or it will be blank.
I have captured some scenarios with this implementation below(I am putting images because it is faster):
Both `$rev1` and `$rev2` out of bounds
image

One of the `$rev1` or `$rev2` out of bounds
image

Both `$rev1` and `$rev2` are under bounds but no changes between revision
image

Both `$rev1` and `$rev2` are under bounds with changes between revision
image

michael-o: Will Subversion tell you that the rev is out of bounds?

k10blogger: What we will get in the log is a null value. If we dump the history variable we get a null value at log entry node. That is how we come to know it is out of bounds.
`object(SVNLog)#5 (3) { ["entries"]=> array(0) { } ["curEntry"]=> NULL ["path"]=> string(14) "/trunk/main.py" }`
In all other scenarios we always get a "currEntry" node.

michael-o: So this is OK for my. Shit it, nothing out.

ams-tschoening: So what's your argument in the end? That printing the warning additionally is NOT necessary in your test cases? Might be, but what's the problem with printing it ADDITIONALLY? Remember that that's exactly how the current implementation works: It stores some warning and proceeds to generate further content. If that content' doesn't make much sense, bad luck, but because of the warning one knows. If that content DOES make sense, good for the user, but the warning makes one have a more detailed look as well. Remember it's a warning only, almost ALL software I know of continues work after logging/showing/... warnings.

Besides that, if one revision is out of bands and you are able to detect that, this qualifies for a warning in my opinion as well. Not only if both revisions are out of band.

k10blogger: Yeah.. My argument was to convert the warning to error. But now reading your argument maybe that is not the best way forward. It really makes no difference to user if he sees warning in addition to content.  
This makes a way better sense than the direction i was thinking in.
Okay i have pushed a change in template where i removed the else condition and added a success check in all the templates.

templates/Elegant/compare.tmpl
```diff
@@ -35,7 +35,8 @@
     </div>
     [websvn-test:warning]
       <div id="warning">[websvn:warning]</div>
-    [websvn-else]
+    [websvn-endtest]
+    [websvn-test:success]
```
ams-tschoening, line 39: Are you sure the success-test is what we want? Look at the BlueGrey-template, it seems to be ONLY used for adding some navigation-block. OTOH, your success-checks in the other templates seem to span ALL content instead. This seems to defeat the purpose from showing content as often as possible. Even if `success` is `false`, `comp.php` continues to work:

				// Check for error messages
				if (strncmp(trim($line), 'svn: ', 5) == 0) {
					$listing[$index++]['info'] = urldecode($line);
					$vars['success'] = false;
					continue;
				}

So I suggest removing `[websvn-test:success]` and the corresponding closing statement again OR make it only work like in `BlueGrey`, for some navigation. But seems easier to me to remove that special treatment from `BlueGrey` as well instead. We don't need to care if that special navigation is hidden sometimes or not in my opinion.

k10blogger: Ahh okay this is an issue with my understanding of the success block. Then i need to remove one of the `[websvn-endtest]` from bottom as well. Let me do that.

Issue #114, opened 2020-08-14
diff should return "no changes in file" between revisions


k10blogger:
Take an example of structure as below:
Say we are at current revision 40
--- SomeFile @ 40
--- TargetFile @ 23
What this means that SomeFile is at revision 40.
TargetFile too is at revision 40 but has not changed since revision 23. 

In scenario if i am doing a comparison using a compare path and put this TargetFile and put the revision as 38 - 40. 
The diff should return `bool(false)` meaning it says that there has been no change in the TargetFile from revision 38-40
This is what is currently missing. It would be good to have this feature. 

As it has been implemented currently:
![image](https://user-images.githubusercontent.com/4973182/90289857-62138e00-de9a-11ea-9617-c2543050651d.png)

We just check that `$diff` stream is returned. We dont check whether the `$diff` stream is returned as false.
A check for that should be added and we should print that "No changes found between commits" or something similar.

ams-tschoening:
When comparing two identical files in e.g. TortoiseSVN, it simply shows their content without any changes. It DOESN'T hide the content and show a message only saying that there's no DIFF. It doesn't even additionally print a message that there's no DIFF.

So how do things look like currently in your scenario and How exactly do you suggest new output to look like? Print the text you mentioned only or additionally or ...? This is not clear to me. I like the approach of TSVN showing the content even in case no changes and that might simply be enhanced by a message like yours above or underneath that text. Should simply be handled like or even as a warning, in which case it makes additional sense to not print warnings exclusively, but additionally.

https://github.com/websvnphp/websvn/pull/113#discussion_r471092911

k10blogger:
Currently it shows blank as this:
![image](https://user-images.githubusercontent.com/4973182/90339624-8d6bb980-e00f-11ea-9375-8b2b0cfcb723.png)
What i am proposing is below:
![image](https://user-images.githubusercontent.com/4973182/90339653-b2602c80-e00f-11ea-8168-b9ed7f355582.png)

This is achieved when the first line ever read from the $diff stream is of length 0.
![image](https://user-images.githubusercontent.com/4973182/90339758-90b37500-e010-11ea-8a95-9e24dd1e0385.png)

So this will be like if or else situation.
If the first line is blank there is no change between revisions even though the commit is being shown. Else there will be changes and warning will never be generated as lines in $diff stream will never be 0.
This can be see by dumping history variables in the debug message.
[...]

PR #115, opened 2020-08-17
diff should return "no changes in file" between revisions for No Changes between revisions.

k10blogger:
This PR is for issue #114 .

Here I have reused the warning variable to show the warning that no changes in file/path have been seen between revisions. 
This warning will be recorded only when no previous warning have been stored in the warning variable. 
The warning is triggered only when the websvn shows that diff between revision is not there.

### DIFF
--- a/comp.php
+++ b/comp.php
@@ -474,6 +474,13 @@ if ($rep) {
 				}
 
 				$listing[$index++]['info'] = escape(toOutputEncoding($line));
+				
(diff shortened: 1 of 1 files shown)


ams-tschoening:
This is a good example that the `warning`-stuff most likely needs to be enhanced at some point as well to allow multiple of those. But the current change is worth to be used of course. Thanks!

PR #121, opened 2020-09-02
Issue 119 PR for File Search using the svn command

k10blogger:
I dropped the work on `svnlookup` and did something less intrusive using the existing `svn` command.
Here is full functioning attempt.

Based on the path that has been browsed the search parameters sets accordingly. Search uses `glob` type search string as in the svn book.


languages/english.php
```diff
@@ -116,3 +116,7 @@
 $lang["WITHPATH"] = "With Path:";
 $lang["FILEDELETED"] = "File deleted";
 $lang["FILEADDED"] = "New file";
+
+$lang['SEARCH'] = "Search";
+$lang['SEARCH_PLACEHOLDER'] = "Search for file name here";
```
michael-o, line 121: Files only? Won't it match directories?

languages/english.php
```diff
@@ -116,3 +116,7 @@
...
+$lang['NOSEARCHTERM'] = "No Search Found";
```
michael-o, line 122: The key does not correspond to the message. It that supposed to be "No Items Found"?

languages/english.php
k10blogger: It matches directories as well. It is just some text written for help. Hoping one of you would let me know what to write here. :)

languages/english.php
k10blogger: Yes it is No Items Found from the search term that was entered.

languages/english.php
michael-o: Make it agnostic: "Search for items here"

languages/english.php
michael-o: The please align key and value.

ams-tschoening (changes requested):
I won't be able to have a deeper look into this any time soon, so proceed as how you see fit. But I strongly suggest having a second look at the code style:

* Newlines
* Braces for `if`
* Early aborts/returns/breaks

All those things I mentioned in #120 already, as those make the code of this PR much more readable.

languages/english.php
k10blogger: Changed the value for key 'SEARCH_PLACEHOLDER' to "Search for items here."

languages/english.php
k10blogger: Changed the key to "NOITEMSFOUND" and value as "No Items Found". If this is okay will do the same for all languages.

search.php
```diff
@@ -61,130 +72,178 @@ function showSearchResults($svnrep, $path, $searchstring, $rev, $peg, $listing,
...
+		$index++;
+		$last_index = $index;
 	}
 
 	return $listing;
 }
+
 // Make sure that we have a repository
-if ($rep) {
+if ($rep)
```
ams-tschoening, line 209: You should change this as well. This is unnecessary bad style in many files currently:

```
if (!$rep)
{
    http_response_code(404);
}

[...]
```

Makes things much more readable.

ams-tschoening (changes requested):

search.php
k10blogger: But this is in an if else case.
```
if ($rep) { --- }
else { http_response_code(404);}
renderTemplate('directory')
```
So if we do this the other section will still be evaluated when it should not. As http_response_code i think only updates the header doesnt exits the code.

ams-tschoening: You are right, I didn't make things explicit I had in mind. My suggestion should really have been the following:

```
if (!$rep)
{
    http_response_code(404);
    renderTemplate('directory');
    exit(0);
}

[...]
renderTemplate('directory');
```

While that duplicates the call to `renderTemplate`, aborting early in my opinion is better to read and maintain. One could even make things easier with something like the following, putting the call to `exit` into the function:

```
function renderTemplate($view, $andExit = false) {
```

Or something like that, handling the whole error case in one call:

```
function renderTemplateNoRep($view)
{
    http_response_code(404);
    renderTemplate($view);
    exit(0);
}
```

That combination of checking for an available repo and print `404` otherwise happens pretty often. At some point the template for that will most likely change as well, because that's an error which would need to be printed specially to the user. Doesn't make too much sense to print a template designed with a repo being available in mind when it's actually not. Introducing such a function would make it easy to find all those places by name.

k10blogger: I have made an exit first implementation. Though i am thinking should we wrap the rest of the code into a function? Though not sure would it help in some way.

ams-tschoening: Functions often help, because they can be documented and stuff like that. But just putting all available code into one without looking if that makes sense as one function, if it makes sense to document the behaviour of that function etc. doesn't help too much. You just would move the code from one place to another, but it would still be a "do everything"-stuff. That doesn't make too much sense, it would only help if you would break the existing code into pieces worth being maintained as individual function per piece, with individual, concrete docs per function etc.

Look at the example I brought with `renderTemplateNoRepo`:

```
http_response_code(404);
renderTemplate('directory');
exit(0);
```

Instead of copying those three lines over and over again in different files, they need to be outsourced into one function used by all files. That function can be documented. maintained at one place only etc. This is different to simply putting available code in one function just to have it as one function, functions need to serve one concrete purpose like rendering things or processing things or retrieving things or checking things or ...

The bottom line is, keep the available large code as is, because putting it in a function wouldn't improve things a lot. If at all, introduce `renderTemplateNoRepo` instead, because the changes you made now need to be made at some point in future in many other files as well.

k10blogger: I made a function in the utils and pushed it in. This can be used in all other templates as well. we can create a backlog item to fix other files with this function as well. 

For the 404 error handle i totally agree with you. I am still not sure why we are printing the entire template that looks weird as seen in image in #126 

> At some point the template for that will most likely change as well, because that's an error which would need to be printed specially to the user. Doesn't make too much sense to print a template designed with a repo being available in mind when it's actually not.

We should discuss the change in template for error 404 in #126.

include/utils.php
```diff
@@ -416,3 +418,16 @@ function tempnamWithCheck($dir, $prefix) {
 
 	return $tmp;
 }
+
+// }}}
+
+// {{{ renderTemplateNoRepo
+
+function renderTemplateNoRepo($view)
```
ams-tschoening, line 426: Any reason you didn't put this into `template.php`, which already contains a `renderTemplate`? I suggest adding it after `renderTemplate` there to keep things in context.

ams-tschoening (changes requested):

include/utils.php
k10blogger: Yeah i dont remember the thought process. Possibly i thought all small utils to be there. Anyways I moved it to template below the renderTemplate function.

languages/german.php
```diff
@@ -116,3 +116,7 @@
 $lang["WITHPATH"] = "Mit Pfad:";
 $lang["FILEDELETED"] = "Datei gelöscht";
 $lang["FILEADDED"] = "Neue Datei";
+
+$lang['SEARCH'] = "suchen";
+$lang['SEARCH_PLACEHOLDER'] = "Suchen Sie hier nach Artikeln";
```
ams-tschoening, line 121: This should be changed to `Suchen Sie hier nach Elementen`. `Artikel` doesn't make too much sense in this context for a native German, because it's more likely to be used as part of an online-shop or alike.

ams-tschoening (changes requested):

ams-tschoening:
This is how a search result looks for me, is that as expected? Especially the files at the top seem strange. I can easily live with that, though, was just wondering.

Bild_2020-09-27_172944

k10blogger:
Yes that is how it is supposed to look. It retains its indentation. I thought it would be good to know it is under some folder. You can still get the complete path from the file.

languages/german.php
k10blogger: Fixed it in the new commit.

ams-tschoening (approved):

k10blogger:
Merging this please raise an issue if anything is found. Thanks for the support.

PR #122, opened 2020-09-13
Reimplementation of GHI #104 to list all dir contents.

ams-tschoening:
I've had a look at #120 and sadly things didn't work for me: The click-handlers were wrong, using spaces to divide parts of paths broke with my test-repo containing dirs with spaces and depending on a proprietary attribute `customaction` didn't feel right as well, when we only wanted to toggle directories in the end. So I changed all of that and am providing current progress here.

Some things are still missing:

- [x] Remove `customaction` everywhere
- [X] Externalize large handler implementations to individual functions
- [X] Improve docs
- [X] Make regular expression safe to prevent injection

Possible improvements:

- [ ] Change icon of lowest directory without any further content to show it's empty? 
  - currently clicking -> no-op
  - no useful icons for all templates
- [x] Somehow detect if to apply new JS at all? 
  - `$config->setLoadAllRepos(true);`
- [ ] Apply sorting on row titles? 
  - `branches` vs. `branches/some_dir` vs. `branches/`, the latter containing a file
  - `branches/bfriedrich/Neu Textdokument[...]` vs. `index.js`

Clipboard01
Clipboard02
Clipboard03
Clipboard04
Clipboard05
Clipboard06

### DIFF
--- a/templates/calm/directory.tmpl
+++ b/templates/calm/directory.tmpl
@@ -92,7 +92,7 @@ e-node=<img src="[websvn:locwebsvnhttp]/templates/calm/images/e-node.png" alt="[
       </thead>
       <tbody>
       [websvn-startlisting]
-      <tr class="row[websvn:rowparity]" valign="middle">
+      <tr class="row[websvn:rowparity]" customaction="closed" title="[websvn:classname]" valign="middle">
          <td class="path" valign="middle">
             [websvn:compare_box]
             [websvn-treenode]
@@ -138,4 +138,11 @@ e-node=<img src="[websvn:locwebsvnhttp]/templates/calm/images/e-node.png" alt="[
   [websvn:compare_endform]
 </div>
 [websvn-endtest]
+<script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/jquery.min.1.9.1.js"></script>
+<script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>
+    <script type="text/javascript">
+    //<![CDATA[
+    collapseAllDir();
+    //]]>
+    </script>
 [websvn-endtest]
--- a/templates/calm/directory.tmpl
+++ b/templates/calm/directory.tmpl
@@ -71,7 +71,7 @@ e-node=<img src="[websvn:locwebsvnhttp]/templates/calm/images/e-node.png" alt="[
 <div id="wrap">
 [websvn:compare_form]
   <div class="table-responsive">
-   <table>
+   <table id="listing">
       <thead>
       <tr align="left" valign="middle">
         <th scope="col" class="path">[lang:PATH]</th>
--- a/templates/calm/directory.tmpl
+++ b/templates/calm/directory.tmpl
@@ -94,8 +94,8 @@ e-node=<img src="[websvn:locwebsvnhttp]/templates/calm/images/e-node.png" alt="[
       [websvn-startlisting]
       <tr class="row[websvn:rowparity]" valign="middle"
             [websvn-test:loadalldir]
-                customaction="close" title="[websvn:classname]"
-            [websvn-endtest] 
+                title="[websvn:classname]"
+            [websvn-endtest]
         >
          <td class="path" valign="middle">
             [websvn:compare_box]
--- a/templates/calm/directory.tmpl
+++ b/templates/calm/directory.tmpl
@@ -145,10 +145,5 @@ e-node=<img src="[websvn:locwebsvnhttp]/templates/calm/images/e-node.png" alt="[
 [websvn-test:loadalldir]
 <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/jquery.min.1.9.1.js"></script>
 <script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>
-    <script type="text/javascript">
-    //<![CDATA[
-    collapseAllDir();
-    //]]>
-    </script>
 [websvn-endtest]
 [websvn-endtest]
--- a/templates/calm/directory.tmpl
+++ b/templates/calm/directory.tmpl
@@ -92,7 +92,11 @@ e-node=<img src="[websvn:locwebsvnhttp]/templates/calm/images/e-node.png" alt="[
       </thead>
       <tbody>
       [websvn-startlisting]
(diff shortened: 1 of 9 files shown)


k10blogger:
There is one issue in this implementation. 
Consider this i have opened a folder
image
Now if i click on a folder to collapse Something like this happens.
image

Meaning if it has a subfolder with files the top folder doesnt collapse fully.

ams-tschoening:
Please make your click-order clear:

1. everything is collapsed
2. clicking `branches` shows `ALP45`
3. clicking `ALP45` shows `.SimulinkProject`, `BuildHere` and `CFiles`
4. clicking `CFiles` shows all files in there
5. clicking `ALP45` HIDES `.SimulinkProject`, `BuildHere` and `CFiles`, but NOT the files within `CFiles`

Is that correct?

If so, I'm pretty sure to know the reason: When clicking on parents, the current implementation simply toggles their direct children. It DOES NOT make all children toggle their children recursively as well. That's why `CFiles` itself gets hidden, because it's a direct children, but the files in that don't, they are NOT a direct children.

Your screenshots describe a not too wrong behaviour, though: When toggling `ALP45`, all contents of that dir should simply be not visible anymore. But the state that `CFiles` and its children were visible might be kept. So when making the contents of `ALP45` visible again, this should show `CFiles` and all its children exactly like in your screenshot, because `CFiles` itself was NOT toggled when toggling `ALP45`.

The problem is that all rows have the same level currently and different paths need to be distinguished by different titles. This makes it impossible to have `ALP45` itself and some HTML-block containing ALL of its children only. If the HTML would be like that, one could simply hide that block with ALL children and the problem you documented in your screenshot wouldn't occur anymore. But the state of `CFiles` with visible children would be kept.

So, two choices: Change the current implementation to trigger toggling by all children-dirs if those dirs get toggles themself. Change the HTML to better model the parent-child-relation ship corresponding to the paths. I guess the former is easier to do.

k10blogger:
Yes that is correct. 
These three variables where the one doing it for me in the JS. Making sure that all children are register correctly. 
```
let strClassCheck = $target.closest("tr").next().attr("title");
let resCheck = strClassCheck.split(" ");
let performAction
```
But if i have understood your implementation correctly. I think with a slight modification of the variable `pattern` we should be able to push these dependent into the directChildren list.
I think i can modify this `body.children(selector).each(function()` add few lines in these and this can work. 
Let me have a go on this implementation. Then we can have another look.

ams-tschoening:
Putting things into `directChildren` which aren't direct children is the wrong approach. In that case things would be shown recursively as well, while we only want to hide recursively, but show really only direct children. We need to distinguish showing between hiding, which is something I though could be avoided. I pushed a different implementation.

k10blogger:
Ahh this is perfect.... This is how it should work.
Any clean up that needs to be done? This is perfect for me.

k10blogger (approved):

k10blogger (approved):

ams-tschoening:
I still need to address at least some of the points in the initial comment, especially creating documented functions instead of the large handler implementation.

I'm not sure if the current implementation is compatible with the old behaviour as well, I think it isn't, but didn't test yet. I guess there's some additional check needed to only do something IF direct children are available currently at all. If not, this should be the old behaviour, but implementing things this way we don't need to care. If children are available, implement the show/hiding/ else don't. No need to care why children are available in theory.

templates/calm/directory.tmpl
```diff
@@ -138,4 +142,8 @@ e-node=<img src="[websvn:locwebsvnhttp]/templates/calm/images/e-node.png" alt="[
   [websvn:compare_endform]
 </div>
 [websvn-endtest]
+[websvn-test:loadalldir]
+<script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/jquery.min.1.9.1.js"></script>
+<script type="text/javascript" src="[websvn:locwebsvnhttp]/javascript/collapsible.js"></script>
+[websvn-endtest]
```
k10blogger, line 148: @ams-tschoening This is where the check for loadalldir is there. If it is set in config file then and only then the JS will be included. Otherwise it will be the same,

k10blogger:
Should we think of limiting the on click handler
`rowParent.on('click',function()` to `rowParent.children('td.path').on('click', function()`
The reason for this is putting this handler on the entire row disables the original behavior of looking at the log,RSS and check out URL. 
To open the file right now we need to right click and open in new tab as left clicking doesn't work. I am okay with this behavior as anyway once loaded i will probably be opening things in new tab. 

For the other points

> Change icon of lowest directory without any further content to show it's empty?

This i can do. If you have any icons it would be easier.

These two
> Externalize large handler implementations to individual functions
> Make regular expression safe to prevent injection

I dont know how to work this. Though i can do the documentation part.


For this
> Apply sorting on row titles?

How would this benefit? Can you explain a bit.

ams-tschoening:
> rowParent.on('click',function() to rowParent.children('td.path').on('click', function()

I narrowed this down to the directory name itself in theory, please have a look if things work properly.

It seems there's still some bug with lower levels of directories and hiding those:

Clipboard01
Clipboard02

ams-tschoening:
> It seems there's still some bug with lower levels of directories and hiding those:

Should be fixed now as well.

k10blogger:
Hi there is still some issue. 
Have a look at the snapshot.
Files under folders
image

If i collapse the top level folder. The open files remain open.
image

If i try to open it back:
image

Then if i open the top of the file folder again:
image

This becomes confusing when many folders are open.
image

Actual of above snapshot
image

k10blogger:
Some description for the above images. 

What works - 
1 - As seen folder and their direct children expansion collapsing works.
2 - In case of `Platform` folder. If the underlying folders are expanded and if i collapse the `Platform` folder then all foilders under it collapses. Which is correct behaviour.

What doesnt work:
1 - If i go to folder above `Platform` and collapse it the open folders under `Platform` remains open. 
2 - This becomes even crazier when multiple sub folders are open and we go to the very top of the parent folder and collapse all. Every open file item remains open. Which is visible in the second last image.

If we collapse the top folder all under it should collapse. - Not Working
But if we open the the top folder only immediate items or folders should show - Working at all levels.

ams-tschoening:
Are you sure to have the latest commit? Your description reads pretty much like what I had fixed:

https://github.com/websvnphp/websvn/pull/122/commits/50a4c0c0fd44668998b0aa3c3f0b74a6b4368a5d

k10blogger:
Yeah sorry my bad. I had to clean out my local and get again. It is working now. 
I see that the clicks to file are also working. Nice. 
Works for me.  If nothing left we can merge?

PR #123, opened 2020-09-15
Fix line breaks not appearing in commit messages in log view.

JamesAtChamSys:
Line break characters were being stripped before being passed to the
XML parser. This meant that commit messages with newlines in were
displayed as one ugly block of text.

I'm not sure, however, if this patch is the best way of fixing this.

### DIFF
--- a/include/svnlook.php
+++ b/include/svnlook.php
@@ -930,7 +930,7 @@ class SVNRepository {
 		xml_set_character_data_handler($xml_parser, $charData);
 
 		for ($i = 0; $i < $linesCnt; ++$i) {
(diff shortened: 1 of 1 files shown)


michael-o:
Can you show how the broken behaviour looks like?

JamesAtChamSys:
These screenshots show the log view before and after this fix:

websvn-bad
websvn-good

ams-tschoening (approved):

michael-o (approved):

ams-tschoening:
Thanks for fixing this!

PR #127, opened 2020-10-04
Issue 126 - Attempt to show 404 with existing template error tag

k10blogger:
As i had some extra time. I took it upon my self to do following changes with this PR.
1 - Implemented the exit first scenario wherever applicable and tried to bring everything as discuss in the coding guidelines. 
(Probably we should make a small wiki page with a coding guidelines)
2 - Showed how the error tag in the templates can be used for showing the 404 error. 

If we can decide upon a 404 error message we can provide support for multiple languages like `lang['404']` instead of a static text here:
image

Further improvements that are seen though not neccessary is that:
1 - There are 34 instances in the entire websvn where http_response_code is called with an error number (404|403|500)
We can take this opportunity to iumprove scenarios as:
image

2 - Basically this would mean extending the `renderTemplateNoRepo` to something more to add these error messages. Like additional arguments.

Let me know your thoughts.

### DIFF
--- a/blame.php
+++ b/blame.php
@@ -31,191 +31,235 @@ require_once 'include/template.php';
 $vars['action'] = $lang['BLAME'];
 
 // Make sure that we have a repository
-if ($rep) {
-	$svnrep = new SVNRepository($rep);
+if (!$rep)
+{
+	renderTemplateNoRepo('blame');
+}
 
-	// If there's no revision info, go to the lastest revision for this path
-	$history = $svnrep->getLog($path, 'HEAD', 1, false, 2, ($path == '/') ? '' : $peg);
-	if (!$history) {
-		unset($vars['error']);
-		$history = $svnrep->getLog($path, '', '', false, 2, ($path == '/') ? '' : $peg);
-		if (!$history) {
-			http_response_code(404);
-			$vars['error'] = $lang['NOPATH'];
+$svnrep = new SVNRepository($rep);
+
+// If there's no revision info, go to the lastest revision for this path
+$history = $svnrep->getLog($path, 'HEAD', 1, false, 2, ($path == '/') ? '' : $peg);
+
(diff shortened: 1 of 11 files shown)


ams-tschoening:
One 404 error message only doesn't make sense, 404 is used for many different error conditions and each of those conditions need one error message. Each of those messages gets 404 status code, so `renderTemplateNoRepo` would become something like `renderTemplate404` with some `cause` or `errorDetail` or alike as argument. Sometimes that cause would be `NO_PATH`, sometimes it would be `NO_REPO` etc. The function would then do `$vars["error"] = $lang["..."]` and render one special error template most likely. That special error template would need to make sure that no controls are available for which the data is not available, e.g. because a repo is missing.

blame.php
```diff
@@ -33,7 +33,7 @@
 // Make sure that we have a repository
 if (!$rep)
 {
-	renderTemplateNoRepo('blame');
+	renderTemplate404('blame',$lang['NOREP']);
```
ams-tschoening, line 36: You should just forward the key `NOREP` and let `renderTemplate404` handle the lookup in `$lang`. This saves typing by the caller and at some point we might not associate an error message only with that key, like is the case currently, but e.g. a special template as well. You already made `$lang` available in `renderTemplate404` anyway, but most likely simply forgot to use it.

    renderTemplate404('blame', 'NOREP');

And:

    $vars['error'] = "WebSVN 404: " . $lang[$errorDetail];

ams-tschoening (changes requested):

blame.php
k10blogger: ahh.. yes yes... this is my bad... I am fixing this...

k10blogger: Done changed this.

ams-tschoening (approved):

PR #129, opened 2020-10-15
Fix path display in log.php when showChanges is disabled

LordAro:
Without `$config->setLogsShowChanges(true);`, the paths on log.php don't get set, due to the svn command getting run without the verbose flag, meaning that the paths never get output.

Before:

image

After:

image

### DIFF
--- a/log.php
+++ b/log.php
@@ -299,7 +299,7 @@ if (!empty($history))
 
 	if ($brev && $erev) 
 	{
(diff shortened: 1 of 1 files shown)


ams-tschoening (approved):

michael-o (approved):
Brilliant, this has been bugging me for quite some time.

PR #132, opened 2020-11-16
Fix undefined index warning in comp.php

LordAro:
Undefined index: warning in /var/www/websvn/comp.php on line 573


ams-tschoening (approved):

michael-o (approved):

k10blogger:
Thanks for the fix.

PR #135, opened 2021-01-02
Attempt to Render Markdown if README.md found on the current path.

k10blogger:
i had some time so i looked into #83 Markdown support. I have added this using a parsedown library. https://parsedown.org
This is a single file. So i added in the local include. But i can remove it if we want to add it to the composer and let it fetch from there  instead. 

It follows the github flavoured markdown. 
https://github.github.com/gfm/ 

This is how it looks when rendered.
in BluGrey
image

in Calm
image

in Elegant
image

If there is no readme.md there will be nothing shown.
Like Github it only looks for "README.md" nothing else.
Though maybe we can make that configurable if we like.


michael-o (changes requested):
Please revert the formatting changes and as well apply the same formatting as done in the other PHP files. This makes it way easier to read the entire code.

ams-tschoening:
Other code using that style has been [introduced in the past as well](https://github.com/websvnphp/websvn/commit/ceffcd2b47bc0876000a2b529245fcada4d8a915#diff-c064d2cdbd2e3541cce965c751c3f68e4f5edefac940722d2b05ddd7518be1dbR492), no need to be too pedantic. Things are perfectly well readable.

michael-o:
That was bad review...

include/svnlook.php
```diff
@@ -861,6 +861,47 @@ function listFileContents($path, $rev = 0, $peg = '') {
...
+		}
+		else {
+			require_once 'include/parsedown.php';
+		}
+		// Capital P and small p to differentiate one used from PEAR or Composer (P) and (p) is from local include
+		$Parsedown = new Parsedown();
+
+		$cmd = $this->svnCommandString('cat', $path.$file, $rev, $peg);
+
+		if ($result = popenCommand($cmd, 'r'))
```
ams-tschoening, line 892: I suggest early aborts again in case of an error. How the rendered result is printed to the browser might easily become lengthier.

```PHP
$cmd = $this->svnCommandString('cat', $path.$file, $rev, $peg);
if (!($result = popenCommand($cmd, 'r')))
{
    return;
} 

echo('<div id="wrap">');
[...]
```

include/svnlook.php
```diff
@@ -861,6 +861,47 @@ function listFileContents($path, $rev = 0, $peg = '') {
...
+			require_once 'include/parsedown.php';
+		}
+		// Capital P and small p to differentiate one used from PEAR or Composer (P) and (p) is from local include
+		$Parsedown = new Parsedown();
+
+		$cmd = $this->svnCommandString('cat', $path.$file, $rev, $peg);
+
+		if ($result = popenCommand($cmd, 'r')) 
+		{
+			echo('<div id="wrap">');
```
ams-tschoening, line 894: Is that `id` used for similar rendered code already. Otherwise it might be a good idea to add a more descriptive CSS-class. By using a distinct class it might be possible for customized skins to simply hide those parts easily. Something like `md-rendered`, `rendered-md` or alike?

include/template.php
```diff
@@ -263,7 +263,9 @@ function parseTags($line, $vars) {
 //
 // Renders the templates for the given view
 
-function renderTemplate($view) {
+function renderTemplate($view,$readmepath=null)
```
ams-tschoening, line 266: Spaces and camel-case make things read better, like in your other function: `$view, $readmePath = null`

include/template.php
```diff
@@ -272,16 +274,30 @@ function renderTemplate($view) {
...
 		executePlainPhpTemplate($vars);
-	} else {
+	}
+	else 
+	{
 		parseTemplate('header.tmpl');
 		flush(); // http://developer.yahoo.com/performance/rules.html#flush
 		parseTemplate($view . '.tmpl');
-		if ($view === 'directory' || $view === 'log') {
+		if ($view === 'directory' || $view === 'log')
```
ams-tschoening, line 288: I suggest an additional newline to distinguish from the former lines. More brackets as well:

> if (($view === 'directory') || ($view === 'log'))

include/template.php
```diff
@@ -272,16 +274,30 @@ function renderTemplate($view) {
...
 		parseTemplate('header.tmpl');
 		flush(); // http://developer.yahoo.com/performance/rules.html#flush
 		parseTemplate($view . '.tmpl');
-		if ($view === 'directory' || $view === 'log') {
+		if ($view === 'directory' || $view === 'log') 
+		{
 			print '<script type="text/javascript" src="'.$locwebsvnhttp.'/javascript/compare-checkboxes.js"></script>';
 		}
+
+		if($readmepath==null)
```
ams-tschoening, line 293: Missing space: `if (`

ams-tschoening (changes requested):

include/parsedown.php
```diff
@@ -0,0 +1,1994 @@
+<?php
```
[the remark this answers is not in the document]
michael-o: Agree with @ams-tschoening. I checked and parsedown at least offers composer support. There seems to be no PEAR channel.

ams-tschoening:
> That was bad review...

I don't think so, changes to the code style of touched code to make things more readable are acceptable. The changes in this PR are really focused as well.

include/parsedown.php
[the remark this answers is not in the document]
k10blogger: Done Removed it from local and updated the composer.json file.

include/svnlook.php
```diff
@@ -861,6 +861,47 @@ function listFileContents($path, $rev = 0, $peg = '') {
...
+	// {{{ listReadmeContents
+	//
+	// Parse the readme.md file
+	function listReadmeContents($path, $rev = 0, $peg = '') {
+		global $config;
+
+		// Make this Configurable ??
+		$file = "README.md";
+
+		if($this->isFile($path.$file) != True)
```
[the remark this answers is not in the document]
k10blogger: Done. Added the space.

include/svnlook.php
k10blogger: Done.

include/template.php
k10blogger: Done.

include/template.php
k10blogger: Done.

include/template.php
k10blogger: Done

include/svnlook.php
k10blogger: Okay yes this is a good idea. Honestly didnt think of this.  I will do this and get back.

include/svnlook.php
```diff
@@ -861,6 +861,47 @@ function listFileContents($path, $rev = 0, $peg = '') {
...
+
+		// Check for the flag for markdown Move this elsewhere?? or better to not added in composer
+		if (!defined('USE_AUTOLOADER')) {
+			require_once 'Parsedown.php';
+		}
+		else {
+			require_once 'include/parsedown.php';
+		}
+		// Capital P and small p to differentiate one used from PEAR or Composer (P) and (p) is from local include
+		$Parsedown = new Parsedown();
```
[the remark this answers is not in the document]
k10blogger: Done used `mdParser`.
My Composer was not working in my development setup for some reason. Started working after a clean install. So had to do this as a workaround. But that is fixed now and I have removed these from here. Only dependent on Composer for now.

include/template.php
ams-tschoening: Please add the spaces as well, like in your other functions:

```
function renderTemplate($view, $readmePath = null)
```

```
if ($readmePath == null)
```

include/template.php
```diff
@@ -272,16 +274,31 @@ function renderTemplate($view) {
...
 		flush(); // http://developer.yahoo.com/performance/rules.html#flush
 		parseTemplate($view . '.tmpl');
-		if ($view === 'directory' || $view === 'log') {
+
+		if (($view === 'directory') || ($view === 'log')) 
+		{
 			print '<script type="text/javascript" src="'.$locwebsvnhttp.'/javascript/compare-checkboxes.js"></script>';
 		}
+
+		if ($readmePath==null)
```
michael-o, line 294: missing whitespace around equals

listing.php
```diff
@@ -631,5 +631,4 @@ function showTreeDir($svnrep, $path, $rev, $peg, $listing)
 }
 
 $vars['restricted'] = !$rep->hasReadAccess($path, false);
-
-renderTemplate('directory');
\ No newline at end of file
+renderTemplate('directory',$path);
```
michael-o, line 634: missing space after comma

michael-o (changes requested):
Additionally to the inline comments: This should be handled just like syntax highlighting. When Parsedown is not availabe, don't do anything. This solution in implies that Parsedown is always avaiable.

ams-tschoening:
Guten Tag Michael Osipov,
am Montag, 4. Januar 2021 um 20:37 schrieben Sie:

> Additionally to the inline comments: This should be handled just
> like syntax highlighting. When Parsedown is not availabe, don't do
> anything. This solution in implies that Parsedown is always avaiable.

From my understanding, ParseDown is always available when composer is
used or else it fails. Making ParseDown optional only make sense if
it's added to the PEAR-docs as well.

https://websvnphp.github.io/docs/install.html#installation

Mit freundlichen Grüßen,

Thorsten Schöning

-- 
Thorsten Schöning
AM-SoFT IT-Service - Bitstore Hameln GmbH

E-Mail:  Thorsten.Schoening@AM-SoFT.de
Web:     http://www.AM-SoFT.de/

Telefon: 05151-  9468-55
Fax:     05151-  9468-88
Mobil:    0178-8 9468-04

Firmensitz: Bitstore IT-Consulting, Frankfurter Allee 285, 10317 Berlin
Steuernummer 037/230/30566, HR 27198, Amtsgericht Potsdam Geschäftsführer Janine Galonska

michael-o:
> 
> Guten Tag Michael Osipov, am Montag, 4. Januar 2021 um 20:37 schrieben Sie:
> Additionally to the inline comments: This should be handled just like syntax highlighting. When Parsedown is not availabe, don't do anything. This solution in implies that Parsedown is always avaiable.
> From my understanding, ParseDown is always available when composer is used or else it fails. Making ParseDown optional only make sense if it's added to the PEAR-docs as well. https://websvnphp.github.io/docs/install.html#installation Mit freundlichen Grüßen, Thorsten Schöning
> […](#)
> -- Thorsten Schöning AM-SoFT IT-Service - Bitstore Hameln GmbH E-Mail: Thorsten.Schoening@AM-SoFT.de Web: http://www.AM-SoFT.de/ Telefon: 05151- 9468-55 Fax: 05151- 9468-88 Mobil: 0178-8 9468-04 Firmensitz: Bitstore IT-Consulting, Frankfurter Allee 285, 10317 Berlin Steuernummer 037/230/30566, HR 27198, Amtsgericht Potsdam Geschäftsführer Janine Galonska

Thanks the point. We don't impose Composer, when someone uses PEAR this will fail, won't it?

include/template.php
k10blogger: Done.

include/template.php
k10blogger: Done.

listing.php
k10blogger: Done.

include/svnlook.php
k10blogger: The "wrap" is an existing id for rendered content.This is something common across all 3 styles. I tried to copy the same style and rename it to `md-render` or something but seems something is always missed out. Formatting goes away. I guess for now i will keep it as id "wrap" only. This wrap has all definition that will be required for this markdown render like table styles and all. I suggest to keep this for now and lets see if some issue arise will tackle then.

k10blogger:
How about something like this
```
if (!defined('USE_AUTOLOADER')) {
	try
	{
		if (!file_exists('Parsedown.php'))
		{
			throw new Exception ("Parsedown.php does not exist");
		}
		else
		{
			require_once 'Parsedown.php';
		}
	}
	catch (Exception $error)
	{
		echo ("Cannot Render README.md as dependent Parsedown library not found. " . $error->getMessage());
		return;
	}
}
```
If library is not found it will just print little message instead of breaking the whole flow.
image

michael-o:
> 
> 
> How about something like this
> 
> ```
> if (!defined('USE_AUTOLOADER')) {
> 	try
> 	{
> 		if (!file_exists('Parsedown.php'))
> 		{
> 			throw new Exception ("Parsedown.php does not exist");
> 		}
> 		else
> 		{
> 			require_once 'Parsedown.php';
> 		}
> 	}
> 	catch (Exception $error)
> 	{
> 		echo ("Cannot Render README.md as dependent Parsedown library not found. " . $error->getMessage());
> 		return;
> 	}
> }
> ```
> 
> If library is not found it will just print little message instead of breaking the whole flow.
> image

This kind of information isn't really relevant to the user. It should be just handled like geshi. If it is not there, fail silently.

k10blogger:
Geshi fails completely if it is enabled and it is not there.

michael-o:
> Geshi fails completely if it is enabled and it is not there.

That's true, but you cannot disable Markdown rendering in config.php.

ams-tschoening:
> How about something like this

Did you actually try your approach? I don't think `file_exists` works the way you think, because when giving a relative path, that is very likely based on the current working directory only. `include_once` OTOH takes special [include paths](https://www.php.net/manual/en/function.include.php) into account, so relative paths for that work differently.

While error messages about missing dependencies are not of interest to the user, WebSVN simply doesn't provide a way to inform admins only about problems. Warnings are printed to the UI in other cases as well, though, simply using `echo` with a HTML-document doesn't make too much sense.

Let's do a step backwards: Which problem should be solved here? Everything contained in `composer.json` is installed ALWAYS, so no need to check anything. That's different when using PEAR, because that is a manual process. Both are different from if some admin WANTS to have markdown rendering enabled at all.

We agreed to not need a config for the file `README.md`, so in my opinion it's fine to NOT need a config to enable/disable markdown rendering at all. IF we want a config to enable/disable, we simply do EXACTLY the same like is done for [GESHI](https://github.com/websvnphp/websvn/blob/4811e65e222c2bbdac058deecc92deb3902db0b8/include/svnlook.php#L762): DO NOT check if files are present or not, if the feature is enabled and `Parsedown` missing, simply break. DISABLE the feature by default to keep backwards compatibility.

```php
if ($this->geshi === null) {
	if (!defined('USE_AUTOLOADER')) {
		require_once 'geshi.php';
	}
	$this->geshi = new GeSHi();
}
```

At some point, PEAR will not be of much use anyway anymore, composer needs to be used and provides all dependencies.

k10blogger:
> Did you actually try your approach? I don't think file_exists works the way you think, because when giving a relative path, that is very likely based on the current working directory only. include_once OTOH takes special include paths into account, so relative paths for that work differently.

Yes relative path is based on current working directory only. But here, there is no expectation of relative path. So it checks for the file in all the included paths as well. For this particular scenario this seems to work. 

The problem statement here is
How to handle this dependency in non composer systems without adding the config variable. Or how to let admins know that this particular dependency is missing. 

I am not that sure how to handle this, i am open to any suggestion. Logical suggestion would be to drop pear support completely(?). I mean most systems now use composer for handling the dependency anyways. 

Any solutions??

ams-tschoening:
> How to handle this dependency in non composer systems without adding the config variable.[...]

1., as we already have them, simply add `Parsedown` to the [PEAR docs](https://github.com/websvnphp/websvnphp.github.io/blob/dcf3cdd87f132cc0a476ecb275e4540a3921fdb1/docs/install.html#L114).

2. change loading of `Parsedown` like is done for GESHI, without any additional checks. This way people using Composer get it, people following the PEAR-docs know about it.

```php
if (!defined('USE_AUTOLOADER')) {
	require_once 'Parsedown.php';
}
```

3. decide if to add a new config variable exactly like is available for GESHI to enable/disable the markdown rendering as a feature, disabled by default. Without such a config, when releasing, one can [easily add a hint](https://github.com/websvnphp/websvn/releases/edit/2.6.0) that `Parsedown` is necessary as well. OTOH, adding such a config shouldn't be too much work for you as well. In the easiest case, just add the config, remember that we only decided against a config for `README.md` in the first place.

michael-o:
I like @ams-tschoening reasoning!

k10blogger:
Thanks for the patience. Got busy someplace else. 
Okay.
So i have added the config for the Mardown. Have a look. 

I have one problem here. The Parsedown library doesnt have a PEAR package at all. I checked at the website also at the pear distribution website but not at all able to run the `PEAR install` command for this Parsedown.

include/distconfig.php
```diff
@@ -410,6 +410,18 @@
 
 // }}}
 
+// {{{ Mardown Render
+
+// Uncomment this line if you want to enable Mardown Rendering of README.md file in the path.
```
ams-tschoening, line 415: Spelling: `Mardown Rendering` vs. `Markdown Rendering`

include/distconfig.php
```diff
@@ -410,6 +410,18 @@
 
 // }}}
 
+// {{{ Mardown Render
+
+// Uncomment this line if you want to enable Mardown Rendering of README.md file in the path. 
+// You will need the Parsedown.php library for this to work. You can either manually add this 
+// library to include path or use the composer to install the same.
+// This will look for REAMDME.md file on the path and render it.
```
ams-tschoening, line 418: Spelling: `REAMDME.md` vs. `README.md`

Suggested improvement:

> This will look for README.md file on the currently navigated path and render it.

include/distconfig.php
```diff
@@ -410,6 +410,18 @@
...
+// The README.md is not configurable for now.
```
ams-tschoening, line 419: Suggested improvement:

> The name of "README.md" isn't configurable for now to simply follow GitHub's conventions.

include/distconfig.php
```diff
@@ -410,6 +410,18 @@
 
 // }}}
 
+// {{{ Mardown Render
+
+// Uncomment this line if you want to enable Mardown Rendering of README.md file in the path. 
+// You will need the Parsedown.php library for this to work. You can either manually add this
```
ams-tschoening, line 416: Suggested improvement:

> You will need the Parsedown.php (https://github.com/erusev/parsedown) library for this to work.

People should now where to get the file from without Composer.

ams-tschoening (changes requested):
Minor changes to the docs, otherwise things look good to me.

include/distconfig.php
k10blogger: Done.

include/distconfig.php
k10blogger: Done.

include/distconfig.php
k10blogger: Done.

include/distconfig.php
k10blogger: Done. Agreed.

ams-tschoening (approved):

include/svnlook.php
```diff
@@ -861,6 +861,50 @@ function listFileContents($path, $rev = 0, $peg = '') {
...
+		{
+			return;
+		}
+
+		if (!$config->getUseParsedown())
+		{
+			return;
+		}
+
+		// Autoloader Handles most of the time
```
michael-o, line 884: handles

include/distconfig.php
```diff
@@ -410,6 +410,17 @@
 
 // }}}
 
+// {{{ Markdown Render
+
+// Uncomment this line if you want to enable Mardown Rendering of README.md file in the path.
```
michael-o, line 415: Typo

michael-o (changes requested):

include/svnlook.php
k10blogger: Changed.

include/distconfig.php
k10blogger: Fixed.

include/svnlook.php
```diff
@@ -861,6 +861,50 @@ function listFileContents($path, $rev = 0, $peg = '') {
 		}
 	}
 
+	// }}}
+	
+	// {{{ listReadmeContents
+	//
+	// Parse the readme.md file
```
michael-o, line 868: Attention, it will not parse `readme.md` on POSIX, but `README.md`.

k10blogger: Oh yes... Corrected now.

michael-o (approved):

k10blogger:
Thanks for the thorough review.

PR #157, opened 2022-02-05
Added line numbers in file detail view with ids for future styling.

k10blogger:
This is a minor enhancement. Previously the view looked like the following:
image

There were no line numbers in the file details view.

The change in the PR adds the line number and provides a basic default styling looks like the following:
image

The change uses the in-built line number system provided by geshi. So viewable only if geshi is used. Same can be done for enscript but I have to read up on it.
Further added an argument highlight to the applyGeshi function to identify file and line highlighting.
The line highlighting number is taken care internally it is only during file highlighting that the line numbers are missing.

### DIFF
--- a/include/svnlook.php
+++ b/include/svnlook.php
@@ -684,7 +684,7 @@ class SVNRepository {
 		$shouldTrimOutput = false;
 		$explodeStr = "\n";
 		if ($highlight != 'no' && $config->useGeshi && $geshiLang = $this->highlightLanguageUsingGeshi($path)) {
-			$this->applyGeshi($path, $tempname, $geshiLang, $rev, $peg);
+			$this->applyGeshi($path, $tempname, $geshiLang, $rev, $peg, $highlight);
 			// Geshi outputs in HTML format, enscript does not
 			$shouldTrimOutput = true;
 			$explodeStr = "<br />";
@@ -780,7 +780,7 @@ class SVNRepository {
 	//
 	// perform syntax highlighting using geshi
 
-	function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = false) {
+	function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = false, $highlight = 'file') {
 		// Output the file to the filename
 		$error	= '';
 		$cmd	= $this->svnCommandString('cat', $path, $rev, $peg).' > '.quote($filename);
@@ -808,8 +808,10 @@ class SVNRepository {
 		$this->geshi->set_header_type(GESHI_HEADER_NONE);
 		$this->geshi->set_overall_class('geshi');
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
-		$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
-		$this->geshi->enable_ids(true);
+		if ($highlight == 'file') {
+			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
+			$this->geshi->enable_ids(true);
+		}
 
 		if ($return) {
 			return $this->geshi->parse_code();
--- a/include/svnlook.php
+++ b/include/svnlook.php
@@ -812,7 +812,8 @@ class SVNRepository {
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
 
 		if ($highlight == 'file') {
-			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
+			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS,1);
+			$this->geshi->set_overall_id('line');
 			$this->geshi->enable_ids(true);
 		}
 
--- a/include/svnlook.php
+++ b/include/svnlook.php
@@ -808,6 +808,8 @@ class SVNRepository {
 		$this->geshi->set_header_type(GESHI_HEADER_NONE);
 		$this->geshi->set_overall_class('geshi');
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
+		$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
+		$this->geshi->enable_ids(true);
 
 		if ($return) {
 			return $this->geshi->parse_code();
--- a/include/svnlook.php
+++ b/include/svnlook.php
@@ -812,8 +812,8 @@ class SVNRepository {
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
 
 		if ($highlight == 'file') {
-			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS,1);
-			$this->geshi->set_overall_id('line');
+			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
+			$this->geshi->set_overall_id('geshi');
 			$this->geshi->enable_ids(true);
 		}
 
--- a/include/svnlook.php
+++ b/include/svnlook.php
@@ -797,17 +797,20 @@ class SVNRepository {
 		}
 
 		$source = file_get_contents($filename);
+
 		if ($this->geshi === null) {
 			if (!defined('USE_AUTOLOADER')) {
 				require_once 'geshi.php';
 			}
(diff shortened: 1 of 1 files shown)


michael-o:
This is fantastatic, I will happily review this next week.

include/svnlook.php
```diff
@@ -808,8 +808,10 @@ function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = fals
 		$this->geshi->set_header_type(GESHI_HEADER_NONE);
 		$this->geshi->set_overall_class('geshi');
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
-		$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
-		$this->geshi->enable_ids(true);
+		if ($highlight == 'file') {
```
ams-tschoening, line 811: Looks good to me, but please add a newline before IF. Makes things more readable. One before line 806 would make sense in my opinion as well.

ams-tschoening (changes requested):

include/svnlook.php
k10blogger: Added newline for readability. At line 800 and 806 as well. Just to be consistent.

ams-tschoening (approved):

include/svnlook.php
```diff
@@ -797,18 +797,25 @@ function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = fals
...
 		}
+
 		$this->geshi->set_source($source);
 		$this->geshi->set_language($language);
 		$this->geshi->set_header_type(GESHI_HEADER_NONE);
 		$this->geshi->set_overall_class('geshi');
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
 
+		if ($highlight == 'file') {
+			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
```
michael-o, line 815: Looking at the docs I do not fully understand fancy here: https://www.wikiparques.com/extensions/SyntaxHighlight_GeSHi/geshi/docs/geshi-doc.html#enabling-line-numbers
What is the second arguement?

michael-o (changes requested):
I see two problems:
1. Anchors aren't clickable, means hard to share
2. Generated ID unusable since it changes constantly, see https://www.wikiparques.com/extensions/SyntaxHighlight_GeSHi/geshi/docs/geshi-doc.html#adding-ids-to-each-line. I guess we need `$geshi->set_overall_id('...')` as well?

k10blogger:
I was wondering how to change the overall ids. Thanks for pointing that out. 
Yes anchors are not clickable but we can jump to anchors as follows(example link):
```
http://localhost/websvn/filedetails.php?repname=Test+Group+1.Common+Model&path=%2Ftrunk%2FCFiles%2Fasm08.c#line-135
```
Using 'line-{line_number}

include/svnlook.php
```diff
@@ -812,7 +812,8 @@ function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = fals
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
 
 		if ($highlight == 'file') {
-			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
+			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS,1);
+			$this->geshi->set_overall_id('line');
```
michael-o, line 816: Attention, this does not only apply to the line, but to the entire geshi rendering.

k10blogger: image
Yes this applies to all lines...

include/svnlook.php
```diff
@@ -797,18 +797,26 @@ function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = fals
...
 		}
+
 		$this->geshi->set_source($source);
 		$this->geshi->set_language($language);
 		$this->geshi->set_header_type(GESHI_HEADER_NONE);
 		$this->geshi->set_overall_class('geshi');
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
 
+		if ($highlight == 'file') {
+			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS,1);
```
michael-o, line 815: The `1` doesn't add any benefit. It will just make every line bold.

include/svnlook.php
michael-o: `line-` is too naive. It will apply to all elements of the Geshi rending, not just lines. Lets make it `geshi` just like the `set_overall_class()`

include/svnlook.php
```diff
@@ -797,18 +797,26 @@ function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = fals
...
+			$this->geshi->set_overall_id('line');
+			$this->geshi->enable_ids(true);
```
michael-o, line 817: `true` is default value, you can drop it and use zero args.

michael-o (changes requested):

include/svnlook.php
k10blogger: By default fancy line numbers are normal line numbers. Adding 1 as an argument makes them "fancy" or "bold". Initially i had not passed any arguments as i was hoping maybe in future we can somehow pass the line number in the url here. So only those line numbers would be marked as bold. Link in case i pass `#line-135` then only line 135 will be marked rest will be normal.

include/svnlook.php
k10blogger: Are you sure?? making geshi would mean if we have to get a url for say line number 110 we have to use anchor as `#geshi-110`. I am thinking line would be more intuitive. like `#line-110` knowing we are trying to jump to line 110. or should we use `l` instead, like done in blame.php ??  
Here is the url in case of blame `http://localhost/websvn/blame.php?repname=Test+Group+1.Common+Model&path=%2Ftrunk%2FCFiles%2Fasm08.c&rev=23#l110`

include/svnlook.php
michael-o: > By default fancy line numbers are normal line numbers. Adding 1 as an argument makes them "fancy" or "bold". Initially i had not passed any arguments as i was hoping maybe in future we can somehow pass the line number in the url here. So only those line numbers would be marked as bold. Link in case i pass `#line-135` then only line 135 will be marked rest will be normal.

This is not what I see. The number means in what step numbers will be bold. I don't see a reavery to make every number bold.

include/svnlook.php
michael-o: Yes, I am sure. Look into `geshi.php` and usage of `overall_id`. It does *not only* apply to numbers. It is all or nothing. As far I can see `l110` is not possible because the template is `{id}-{number}`. The hyphen remains.

include/svnlook.php
k10blogger: ahh okay. Anyways That is a topic for another PR. Completely dropped the 2nd argument.

include/svnlook.php
k10blogger: Okay... Changed to `geshi`.

include/svnlook.php
```diff
@@ -812,8 +812,8 @@ function applyGeshi($path, $filename, $language, $rev, $peg = '', $return = fals
 		$this->geshi->set_tab_width($this->repConfig->getExpandTabsBy());
 
 		if ($highlight == 'file') {
-			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS,1);
-			$this->geshi->set_overall_id('line');
+			$this->geshi->enable_line_numbers(GESHI_FANCY_LINE_NUMBERS);
+			$this->geshi->set_overall_id('geshi');
```
michael-o, line 816: I don't see a better way without fixing geshi first. The issue that Geshi can be used multiple times in a PHP script, mean you would produce clashing its. That is why you need an overall id.
Bascially, they need to introduce a line number override and we make sure that only a single instance of geshi is used per svnlook. Which is.

michael-o (commented):
Please squash

k10blogger:
@michael-o Unfortunately i did a squash in web view which merged all the commits and changes to the base branch. Please have a look here. If needed i can revert and add changes again.

michael-o:
> @michael-o Unfortunately i did a squash in web view which merged all the commits and changes to the base branch. Please have a look here. If needed i can revert and add changes again.

Not necessary, take it as lessons learned.

PR #192, opened 2023-09-30
Fix "Call to a member function logsShowChanges() on null" on log.php

danielmarschall:
Fixes the following bug:
Call 
.../websvn/log.php?repname=xxx&path=%2F
where xxx is a non-existing repo
Results in Fatal error: Uncaught Error: Call to a member function logsShowChanges() on null in .../log.php:41


michael-o (approved):
Good catch!

PR #212, opened 2024-01-28
New methods have been added to configuration to clear set the encodings of user files.

raspopov:
New methods **setCharacterEncodings(array())** and **getCharacterEncodings()** have been added to configuration.
They allow to set the encodings of user files in a standard way.

For example, for Russian encoding, add the following line to **config.php**:
`$config->setCharacterEncodings(array( 'UTF-8', 'windows-1251' ));`

### DIFF
--- a/include/command.php
+++ b/include/command.php
@@ -23,7 +23,9 @@
 // External command handling
 
 function detectCharacterEncoding($str) {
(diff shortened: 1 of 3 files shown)


michael-o (commented):
LGTM

